PR #2, opened 2016-06-14
New implementation based on julia-0.5 infrastructure

timholy:
https://github.com/JuliaLang/julia/pull/16260 added support for arrays with unconventional indexing. This essentially rewrites this package based on that infrastructure. It seems to be pretty compatible with your implementation; of the examples, only "main_sub" fails, and as far as I can tell this implementation doesn't actually use this package (it uses tricks with `SubArray`, but on julia-0.5 those tricks won't work anyway because bounds-checking has been introduced into `SubArray`s). You could presumably rewrite that example?

I noticed that despite the fact that this is a registered package, there is no tagged version. That means users can't `Pkg.add("OffsetArrays")`. Given the requirement for julia-0.5 here, my recommendation is the following:
- first, before doing anything else you tag an official version of this this packages as it is now, before merging this. It could be tagged something like `v0.1.0` (`Pkg.tag("OffsetArrays", :minor); Pkg.publish()`. You should do that even if you have no intention of merging this.
- create and push a new branch, called `pre-arraymageddon` or something. This is for allowing you to update the old code in case you discover bugs for prior Julia versions.
- then, if/when you're comfortable with this implementation, you can merge this
- tag a version that's a minor-version bump (e.g., `v"0.2.0"`) for the new implementation.

Users not running julia-0.5 will get the old one, and the fact that there's a minor version bump means that you can continue to release new versions for older versions of Julia, basing the changes off the `pre-arraymageddon` branch.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,1 @@
-julia 0.2-
+julia 0.5-
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -1,128 +1,97 @@
 module OffsetArrays
 
-#import Base: Array
-importall Base
+Base.@deprecate_binding (..) Colon()
 
-export OffsetArray, ..
+using Base: SimIdx, Indices, LinearSlow, LinearFast
 
-type OffsetArray{T<:Number, N, A<:AbstractArray} <: AbstractArray
+export OffsetArray
 
+immutable OffsetArray{T,N,AA<:AbstractArray} <: AbstractArray{T,N}
+    parent::AA
     offsets::NTuple{N,Int}
-    array::A
-    o1::Int
-    o2::Int
-    o3::Int
-    o4::Int
-    o5::Int
+end
+typealias OffsetVector{T,AA<:AbstractArray} OffsetArray{T,1,AA}
 
-    function OffsetArray(r::Range1{Int})
-        array = Array(T, length(r))
-        offs = (1 - minimum(r),)
-        new(offs, array, offs[1])
-    end
+OffsetArray{T,N}(A::AbstractArray{T,N}, offsets::NTuple{N,Int}) = OffsetArray{T,N,typeof(A)}(A, offsets)
+OffsetArray{T,N}(A::AbstractArray{T,N}, offsets::Vararg{Int,N}) = OffsetArray(A, offsets)
 
-    function OffsetArray(r1::Range1{Int}, r2::Range1{Int})
-        dims = (length(r1), length(r2))
-        array = Array(T, dims)
-        offs = (1 - minimum(r1), 1 - minimum(r2))
-        new(offs, array, offs[1], offs[2])
-    end
+(::Type{OffsetArray{T,N}}){T,N}(inds::Indices{N}) = OffsetArray{T,N,Array{T,N}}(Array{T,N}(map(Base.dimlength, inds)), map(indsoffset, inds))
+(::Type{OffsetArray{T}}){T,N}(inds::Indices{N}) = OffsetArray{T,N}(inds)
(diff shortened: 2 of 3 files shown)


timholy:
Now that there are tests, you also might want to consider visiting travis-ci.org and turning on Travis testing.

alsam:
@timholy thanks a lot for excellent job and for detailed merging instructions!
I will look into it soon.

alsam:
@timholy thanks again. I merged your changes.
Tried to to tag, but unfortunately get
```julia
julia> Pkg.tag("OffsetArrays", :minor)
ERROR: Pkg.tag(pkg, [ver, [commit]]) has been moved to the package PkgDev.jl.
Run Pkg.add("PkgDev") to install PkgDev on Julia v0.5-
 in tag(::String, ::Symbol) at ./pkg/pkg.jl:263
 in eval(::Module, ::Any) at ./boot.jl:231
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```
despite the fact I followed the above recommendations and made
```julia
julia> Pkg.add("PkgDev")
INFO: Cloning cache of PkgDev from https://github.com/JuliaLang/PkgDev.jl.git
INFO: Installing PkgDev v0.1.0
INFO: Package database updated
```
the error message for `ERROR: Pkg.tag` stuck.
So unfortunately the package left untagged. 
Will look into it later.

alsam:
Ok, looks like figured out what went wrong.
Going to do the following: revert @timholy changes, tag correctly, publish, after that merge the changes again with the correct tag for Julia 0.5.

timholy:
Sorry I missed this conversation (buried below some other emails), but glad you figured it out!

alsam:
Sure no problem, and thanks again for your contribution!

I've already tagged with v0.1.3 the old pre-0.5 code. I put it to `pre-arraymageddon` branch, as you proposed and after hat pushed your code to `master`.
However something is broken with the new `PkgDev` package:
tagged it ok with
```julia
julia> PkgDev.tag("OffsetArrays", v"0.2.0");
INFO: Tagging OffsetArrays v0.2.0
INFO: Committing METADATA for OffsetArrays
```
but wasn't able to publish the changes:
```
julia> PkgDev.publish()
...
INFO: Validating METADATA
INFO: Pushing OffsetArrays permanent tags: v0.2.0
ERROR: MethodError: no method matching Base.LibGit2.UserPasswordCredentials(::Void, ::String)
Closest candidates are:
  Base.LibGit2.UserPasswordCredentials(::AbstractString, ::AbstractString)
  Base.LibGit2.UserPasswordCredentials{T}(::Any)
 in credentials() at .../.julia/v0.5/PkgDev/src/github.jl:126
 in (::PkgDev.Entry.##6#9{Dict{String,Array{String,1}}})(::Base.LibGit2.GitRepo) at .../.julia/v0.5/PkgDev/src/entry.jl:99
 in with(::PkgDev.Entry.##6#9{Dict{String,Array{String,1}}}, ::Base.LibGit2.GitRepo) at ./libgit2/types.jl:660
 in publish(::String, ::String) at .../.julia/v0.5/PkgDev/src/entry.jl:84
 in publish() at .../.julia/v0.5/PkgDev/src/PkgDev.jl:55
 in eval(::Module, ::Any) at ./boot.jl:231
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```

timholy:
Yep, that's perhaps the single biggest release blocker for julia-0.5.

alsam:
I see, I filed an issue for `PkgDev`
https://github.com/JuliaLang/PkgDev.jl/issues/48

PR #3, opened 2016-07-27
Update for recent changes in Base

timholy:
This is consistent with http://docs.julialang.org/en/latest/devdocs/offset-arrays/ and with the test code in https://github.com/JuliaLang/julia/blob/master/test/offsetarray.jl.


alsam:
Thanks @timholy !

PR #5, opened 2016-08-12
Fixes and track changes in base

timholy:


alsam:
Thanks @timholy ! Noticed in travis log:
```sh
ERROR: LoadError: size not supported for arrays with indices (0:4,1:8); see http://docs.julialang.org/en/latest/devdocs/offset-arrays/
```
noticed in the docs above:
```sh
To ensure that such errors are caught, in Julia 0.5 both length and size should throw an error when passed an array with non-1 indexing. This is designed to force users of such arrays to check the code, and inspect it for whether it needs to be generalized.
```
Ok to fix it myself?

timholy:
Oh, it looks like the FFT changes haven't been backported to 0.5 yet. See https://github.com/JuliaLang/julia/pull/17919

alsam:
Ah, I see. Thanks!

timholy:
Looks like they didn't make it into rc2 either. I could drop that last commit if you prefer, and add it back in once it's been backported.

alsam:
No problem. I'm going to do the following:
1. merge your changes.
2. revert the last commit.
3. tag and create pull request for METADATA without last commit.
4. I could add the reverted commit later after the changes have been backported - or you what would you prefer.
ok to go ahead?

timholy:
Sounds great to me.

alsam:
Unfortunately I'm not able to tag correctly and publish the new version.
Suspect a bug in `PkgDev`. Details are here:
https://github.com/JuliaLang/METADATA.jl/pull/5939/files#r74563796

PR #6, opened 2016-08-17
reshaping an OffsetArray reshapes the parent

timholy:
When reshaping there's no need to keep the original OffsetArray container, it's possible this fixes a bug (I can't remember, I've had this for several days now).

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -64,6 +64,7 @@ Base.similar(f::Union{Function,DataType}, shape::Tuple{UnitRange,Vararg{UnitRang
 
 Base.reshape(A::AbstractArray, inds::Tuple{UnitRange,Vararg{UnitRange}}) = OffsetArray(reshape(A, map(length, inds)), map(indexoffset, inds))
 
+Base.reshape(A::OffsetArray, inds::Tuple{UnitRange,Vararg{UnitRange}}) = OffsetArray(reshape(parent(A), map(length, inds)), map(indexoffset, inds))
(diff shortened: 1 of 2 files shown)


alsam:
Thank you for the patch.

alsam:
@timholy I tagged correctly your patch with `v0.2.4` . It points to `fbed71ff228afd4298`.
Unfortunately `PkgDev` is broken for me even worse as the previous workaround doesn't work.
I'd rather create a branch of `METADATA` for push request for the new tag and open the request from `github` webpage manually, as `sha1` is ready and look good for me. Ok to do it?

timholy:
That seems reasonable. If you like, I could also do the tagging if you add me as a collaborator. Totally up to you, of course.

timholy:
I'm sorry git is so troublesome. I remember suffering for months until it became second nature. Difficulties in making the switch to git put the final nail in the coffin of my involvement in my first open-source project, [tuxmath](https://tux4kids.alioth.debian.org/tuxmath/). Of course, the bigger reason was that my kids had gotten past the stage of needing help practicing their arithmetic, so I was losing interest anyway :wink:.

alsam:
I've added you as collaborator. Would rather finish tagging with `v0.2.4` myself  - it is a good exercise :-)

alsam:
Speaking about [tuxmath](https://tux4kids.alioth.debian.org/tuxmath/) my daughter used to play in it.

timholy:
Cool! Always nice to hear that others got value out of it.

And thanks for the tag!

PR #9, opened 2016-08-20
Faster benchmarks

timholy:
Thanks for adding the benchmarks! I noticed I could make them even faster (in some cases, from ~300microseconds down to ~4microseconds) by getting rid of the global variables.

It's satisfying that we're fairly close to the performance of `Array`s. One of the things to note is that these benchmarks don't use `@inbounds`; if they did, this would look worse. For Julia 0.6, we'll be able to elide bounds checking and catch up, but that's too dangerous for 0.5 (it would generate a lot of segfaults when used in other packages that haven't been updated for non-1 indexing).

I also made `similar` a little faster; that doesn't have anything to do with these benchmarks, though.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -6,17 +6,13 @@ const dim = 1000
 x = Array(Float64, 2*dim)
 y = OffsetArray(Float64, -dim + 1 : dim)
 
-fillx() = for i = 1:2*dim x[i] = i end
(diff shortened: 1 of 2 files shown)


alsam:
Thanks for faster benchmarks! I've also noticed that the performance of `OffsetArray`
for `v0.5` and `v0.6` due to your contributions is quite competitive
and pretty close to the performance of `Array`. Thanks!

PR #10, opened 2016-08-24
Add unsafe macro to eliminate bounds checking

timholy:
This PR deserves careful thought, and it's possible we shouldn't merge it. We've made a deliberate decision to recommend against making `@inbounds` work with arrays like OffsetArrays, because there's too much code in the wild that has marked sections of code with `@inbounds` for which it isn't actually safe unless the indices start at 1.

However, in practical usage the lack of bounds-check elimination can be a big deal; in our `update2d` benchmark, there's about a 3-fold performance gap between OffsetArrays and Arrays. In applications like [ImagesFiltering.jl](https://github.com/JuliaImages/ImagesFiltering.jl/tree/teh/finish2) where I'm making extensive use of OffsetArrays, this is noticeable enough that I found myself starting to write special implementations for OffsetArrays that stripped off the container; after a while I realized that was stupid and did this instead.

The hack here is to use a different name, here called `@unsafe`, which accomplishes something similar to `@inbounds` (in a much less sophisticated/powerful way, but good enough for quite a few cases). The point being that since there isn't any code out there already using `@unsafe`, perhaps this is OK. In the new benchmarks using `@unsafe`, OffsetArrays match the performance of Arrays.

Thoughts? CC @JeffBezanson to give him the chance to disagree.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -9,9 +9,9 @@ x2d = Array(Float64, 2*dim, 2*dim)
 y2d = OffsetArray(Float64, -dim + 1 : dim, -dim + 1 : dim)
 
 fill(x) = for i in indices(x,1); x[i] = i; end
-fill2d(x) = for i in indices(x,1); for j in indices(x,2); x[i,j] = i + j; end; end
+fill2d(x) = for j in indices(x,2); for i in indices(x,1); x[i,j] = i + j; end; end
 update(x) = for i in indices(x,1); x[i] = x[i] + i; end
-update2d(x) = for i in indices(x,1); for j in indices(x,2); x[i,j] = x[i,j] + i + j; end; end
+update2d(x) = for j in indices(x,2); for i in indices(x,1); x[i,j] = x[i,j] + i + j; end; end
 update_eachindex(x) = for i in eachindex(x); x[i] = x[i] + i; end
 
 @show @benchmark fill(x)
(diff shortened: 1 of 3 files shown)


alsam:
Thanks @timholy ! This macro is really great. BTW thanks for fixing indices ordering in `update2d` - I'm used to C-like row-major ordering but Julia as matlab and Fortran has column-major 2d arrays ordering - I should have remembered it. In my opinion it is too costly to pay such a big performance penalty for bounds checking. Some time ago I noticed, if comment `checkbounds` within `getindex`, `setindex` it will be possible to get a drastic performance boost for `OffsetArrays`.
Ideally it would be nice to have an opportunity at `julia` compiler invocation level to have a switch for selectively suppress bounds checking, e.g. for `OffsetArrays` only as suppressing it everywhere generally is a bad idea and in many places the penalty is not so high, say 15%. Summing up ideally it would be great to have the same code and the opportunity to switch off bounds checking for `OffsetArrays` only using some special julia flag (maybe environment variable?). Anyway I'm for merging. Thanks for the patch.

timholy:
Eventually we can just add `@boundscheck` in front of the `checkbounds` statements, and then `@inbounds` will eliminate the checks. But that's too dangerous to do before people have a chance to get nice error messages (rather than segfaults) when passing `OffsetArray`s to package code that doesn't expect indexing to start at something different from 1.

PR #12, opened 2016-08-26
Deprecate zeros(inds::UnitRange...) and similar `ones`

timholy:
Sadly, I think we need to eliminate these methods. The problem is that a `UnitRange` is an `AbstractArray`, and `zeros([T], A::AbstractArray)` already means something different (create an all-zeros array of element type `T` and size of `A`). So making a 1-dimensional OffsetArray this way is ambiguous. Sorry I didn't notice this conflict earlier.

I made them a deprecation rather than an error in case anyone's code is relying on these. But at a certain point we should delete the deprecations.


alsam:
That's fine, thank you for noticing the conflict and for the patch. Merging.

PR #19, opened 2017-02-11
Fixes for Julia 0.6

timholy:
There are still some deprecation warnings (generated by the tests, not the code in `src`). This requires https://github.com/JuliaLang/julia/pull/20573 to pass tests on 0.6, but makes sense even without it.

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -83,12 +83,12 @@ end
     @inbounds ret = parent(A)[offset(A.offsets, I)...]
     ret
 end
-@inline function Base._getindex(::LinearFast, A::OffsetVector, i::Int)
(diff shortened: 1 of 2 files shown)


timholy:
See also #20. Might be cleaner to take that one for the `similar` and then I can rebase this. CC @pabloferz.

pabloferz:
I missed that you already had something here. You can bring the change in #20 if you want and I'll close that one.

timholy:
No, I like this a little better. Let's merge it.

timholy:
Rebased.

timholy:
I don't see any need to wait for resolution on JuliaLang/julia#20573, so I'll merge and tag a new release.

timholy:
@alsam, you might consider configuring [attobot](https://github.com/attobot/attobot), it makes releases much easier.

PR #20, opened 2017-02-11
Fix similar to take UnionAll into account

pabloferz:
x-ref https://github.com/JuliaLang/julia/issues/20569


timholy:
Duh, right. I didn't think it would matter, but obviously this is better. See also #19.

PR #23, opened 2017-02-18
Fix IndexStyle deprecation on Julia 0.6

timholy:


timholy:
Test failure on 0.6 is because we need at least 0.6.0-dev.2840 and nightly is currently on 0.6.0-dev.2635.

PR #29, opened 2017-09-13
Bump minimum Julia version to fix deprecations

timholy:


alsam:
@timholy thanks a lot for the fixes, i was pretty busy to look into the issue.

PR #32, opened 2017-09-25
Inference improvements on 0.7 and improvements in `show`

timholy:
This does two things:
- fix an inference problem (will require https://github.com/JuliaLang/julia/pull/23867 to pass tests)
- leverage the new `showarg` functionality in Base to make `summary` easier to read. This was prototyped in ImageCore over the past 8 months; with it now in Base, it's time to migrate it into the respective packages.


timholy:
I'll merge this since https://github.com/JuliaLang/julia/pull/23867 has now merged. The Linux nightlies are about 3 weeks behind, so the tests will continue to fail until they catch up.

PR #34, opened 2017-12-02
deprecate constructors with type as first argument

fredrikekre:
This type of `Array` constructors where deprecated in https://github.com/JuliaLang/julia/pull/19989, so `OffsetArrays` should do the same. Including the comma and the space, the replacement is even the same number of characters :smile:

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -33,12 +33,14 @@ OffsetArray{T}(inds::Indices{N}) where {T,N} = OffsetArray{T,N}(inds)
 OffsetArray{T,N}(inds::Vararg{AbstractUnitRange,N}) where {T,N} = OffsetArray{T,N}(inds)
 OffsetArray{T}(inds::Vararg{AbstractUnitRange,N}) where {T,N} = OffsetArray{T,N}(inds)
 OffsetArray(A::AbstractArray{T,0}) where {T} = OffsetArray{T,0,typeof(A)}(A, ())
-OffsetArray(::Type{T}, inds::Vararg{UnitRange{Int},N}) where {T,N} = OffsetArray{T,N}(inds)
 
 # OffsetVector constructors
 OffsetVector(A::AbstractVector, offset) = OffsetArray(A, offset)
-OffsetVector(::Type{T}, inds::AbstractUnitRange) where {T} = OffsetArray{T,1}(inds)
 OffsetVector{T}(inds::AbstractUnitRange) where {T} = OffsetArray{T}(inds)
+
+# https://github.com/JuliaLang/julia/pull/19989
+Base.@deprecate OffsetArray(::Type{T}, inds::Vararg{UnitRange{Int},N}) where {T,N} OffsetArray{T}(inds)
+Base.@deprecate OffsetVector(::Type{T}, inds::AbstractUnitRange) where {T} OffsetVector{T}(inds)
 
 # The next two are necessary for ambiguity resolution. Really, the
 # second method should not be necessary.
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ like Fortran.
 ```julia
 julia> using OffsetArrays
 
-julia> y = OffsetArray(Float64, -1:1, -7:7, -128:512, -5:5, -1:1, -3:3, -2:2, -1:1);
+julia> y = OffsetArray{Float64}(-1:1, -7:7, -128:512, -5:5, -1:1, -3:3, -2:2, -1:1);
 
 julia> summary(y)
 "OffsetArrays.OffsetArray{Float64,8,Array{Float64,8}} with indices -1:1×-7:7×-128:512×-5:5×-1:1×-3:3×-2:2×-1:1"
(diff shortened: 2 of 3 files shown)


fredrikekre:
Tests fail because of https://github.com/JuliaLang/Compat.jl/pull/416#issuecomment-348452332, so need to wait for Compat v0.39.

src/OffsetArrays.jl
```diff
@@ -84,8 +86,10 @@ function Base.similar(A::AbstractArray, ::Type{T}, inds::Tuple{UnitRange,Vararg{
     OffsetArray(B, map(indexoffset, inds))
 end
 
-Base.similar(f::Union{Function,Type}, shape::Tuple{UnitRange,Vararg{UnitRange}}) =
+Base.similar(f::Function, shape::Tuple{UnitRange,Vararg{UnitRange}}) =
     OffsetArray(f(map(length, shape)), map(indexoffset, shape))
+Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:OffsetArray} =
+    OffsetArray(T(map(length, shape)), map(indexoffset, shape))
```
fredrikekre: I had to split this into two methods to fix this on julia master:
```
julia> Test.detect_ambiguities(OffsetArrays, Base, Core)
Skipping Base.Date
2-element Array{Tuple{Method,Method},1}:
 (similar(f::Union{Function, Type}, shape::Tuple{UnitRange,Vararg{UnitRange,N} where N}) in OffsetArrays at /Users/Fredrik/.julia/v0.7/OffsetArrays/src/OffsetArrays.jl:89, similar(::Type{T}, shape::Tuple) where T<:BitArray in Base at bitarray.jl:355)
 (similar(f::Union{Function, Type}, shape::Tuple{UnitRange,Vararg{UnitRange,N} where N}) in OffsetArrays at /Users/Fredrik/.julia/v0.7/OffsetArrays/src/OffsetArrays.jl:89, similar(::Type{T}, shape::Tuple) where T<:Array in Base at array.jl:254)
```

timholy: Will the `T` version trigger your deprecation warning?

fredrikekre:
Yes there is some more stuff to work out here, see e.g. https://github.com/JuliaLang/julia/pull/24785#issuecomment-348704775

I'm planning to take a look tonight.

src/OffsetArrays.jl
```diff
@@ -84,8 +86,14 @@ function Base.similar(A::AbstractArray, ::Type{T}, inds::Tuple{UnitRange,Vararg{
...
+Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:Array} =
+    OffsetArray(T(map(length, shape)), map(indexoffset, shape))
+Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:BitArray} =
+    OffsetArray(T(map(length, shape)), map(indexoffset, shape))
```
fredrikekre: These two `similar` methods make me a bit uncomfortable.

fredrikekre:
Tests pass locally with https://github.com/JuliaLang/julia/pull/24918 now, but not sure about those `similar` methods.

src/OffsetArrays.jl
timholy: What motivated their addition? Presumably a test failure?

fredrikekre:
Rebased after #37

src/OffsetArrays.jl
```diff
@@ -33,13 +33,15 @@ OffsetArray{T}(inds::Indices{N}) where {T,N} = OffsetArray{T,N}(inds)
...
 OffsetArray(A::AbstractArray{T,0}) where {T} = OffsetArray{T,0,typeof(A)}(A, ())
-OffsetArray(::Type{T}, inds::Vararg{UnitRange{Int},N}) where {T,N} = OffsetArray{T,N}(inds)
 
 # OffsetVector constructors
 OffsetVector(A::AbstractVector, offset) = OffsetArray(A, offset)
-OffsetVector(::Type{T}, inds::AbstractUnitRange) where {T} = OffsetArray{T,1}(inds)
 OffsetVector{T}(inds::AbstractUnitRange) where {T} = OffsetArray{T}(inds)
 
+# https://github.com/JuliaLang/julia/pull/19989
+Base.@deprecate OffsetArray(::Type{T}, inds::Vararg{UnitRange{Int},N}) where {T,N} OffsetArray{T}(inds)
```
timholy: What about switching to `uninitialized` syntax now? We're moving now to `ArrayType(iterator)` constructs the array from the contents of `iterator`, but since `-3:3` is an iterator it's ambiguous whether that's supposed to be treated as the initializing values or the axes.

fredrikekre: I have that change as another PR in the pipeline, I can include it in this one if you prefer?

timholy: Either way is fine with me. I'll merge this and you can do that in a next one.

PR #37, opened 2017-12-05
[WIP] fixes for julia 0.7

fredrikekre:
Split out from #34: This PR aims to just get things working on julia master, and then I will rebase #34 on top of this.

So after investigating a bit further regarding the `similiar` methods discussed in #34 there seem to be two problems:

- With the new broadcasting interface, `broadcast_similar` used the fallback method. I fixed this by hooking into the broadcast machinery (excellent interface btw Tim :tada:)

- The other problems show up in the tests I commented out for now. For instance:
    ```julia
    julia> A = OffsetArray(rand(4,4), (-3,5));
    
    julia> unique(A, 1)
    ERROR: MethodError: no method matching Array{Int64,N} where N(::Uninitialized, ::Tuple{UnitRange{Int64}})
    Stacktrace:
     [1] similar(::Type{Array{Int64,N} where N}, ::Tuple{UnitRange{Int64}}) at ./array.jl:254
     [2] similar(::Type, ::UnitRange{Int64}) at ./abstractarray.jl:567
     [3] @generated body at ./multidimensional.jl:1583 [inlined]
     [4] unique(::OffsetArray{Float64,2,Array{Float64,2}}, ::Int64) at ./multidimensional.jl:1571
     [5] top-level scope
    ```

which is because `unique` calls the `similar(::Type{BitArray}, inds)` method here: https://github.com/JuliaLang/julia/blob/c697415e5d665e2085846dc3a3313cff296ef98f/base/multidimensional.jl#L1606 which forces us to hijack that method.

Similarly, for `sortcols` / `sortrows` we end up here: https://github.com/JuliaLang/julia/blob/c697415e5d665e2085846dc3a3313cff296ef98f/base/sort.jl#L964 which forces us to hijack that method. `sortcols` / `sortrows` can be fixed, *I think* with the following patch to base:
```diff
diff --git a/base/sort.jl b/base/sort.jl
index 2c7a0ac..ca7bbf6 100644
--- a/base/sort.jl
+++ b/base/sort.jl
@@ -922,7 +922,7 @@ julia> sortrows([7 3 5; -1 6 4; 9 -2 8], rev=true)
 function sortrows(A::AbstractMatrix; kws...)
     inds = indices(A,1)
     T = slicetypeof(A, inds, :)
-    rows = similar(Vector{T}, indices(A, 1))
+    rows = similar(A, T, indices(A, 1))
     for i in inds
         rows[i] = view(A, i, :)
     end
@@ -961,7 +961,7 @@ julia> sortcols([7 3 5; 6 -1 -4; 9 -2 8], rev=true)
 function sortcols(A::AbstractMatrix; kws...)
     inds = indices(A,2)
     T = slicetypeof(A, :, inds)
-    cols = similar(Vector{T}, indices(A, 2))
+    cols = similar(A, T, indices(A, 2))
     for i in inds
         cols[i] = view(A, :, i)
     end
```

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -4,6 +4,7 @@ module OffsetArrays
 
 using Base: Indices, tail
 using Compat
+using Compat: axes, CartesianIndices
 
 export OffsetArray, OffsetVector, @unsafe
 
@@ -27,7 +28,7 @@ OffsetArray(A::AbstractArray{T,N}, offsets::Vararg{Int,N}) where {T,N} =
     OffsetArray(A, offsets)
 
 OffsetArray{T,N}(inds::Indices{N}) where {T,N} =
-    OffsetArray{T,N,Array{T,N}}(Array{T,N}(map(length, inds)), map(indexoffset, inds))
+    OffsetArray{T,N,Array{T,N}}(Array{T,N}(uninitialized, map(length, inds)), map(indexoffset, inds))
 OffsetArray{T}(inds::Indices{N}) where {T,N} = OffsetArray{T,N}(inds)
 OffsetArray{T,N}(inds::Vararg{AbstractUnitRange,N}) where {T,N} = OffsetArray{T,N}(inds)
 OffsetArray{T}(inds::Vararg{AbstractUnitRange,N}) where {T,N} = OffsetArray{T,N}(inds)
@@ -44,9 +45,9 @@ OffsetVector{T}(inds::AbstractUnitRange) where {T} = OffsetArray{T}(inds)
 OffsetArray(A::AbstractArray{T,0}, inds::Tuple{}) where {T} = OffsetArray{T,0,typeof(A)}(A, ())
 OffsetArray(A::AbstractArray{T,N}, inds::Tuple{}) where {T,N} = error("this should never be called")
 function OffsetArray(A::AbstractArray{T,N}, inds::NTuple{N,AbstractUnitRange}) where {T,N}
-    lA = map(length, indices(A))
+    lA = map(length, axes(A))
     lI = map(length, inds)
-    lA == lI || throw(DimensionMismatch("supplied indices do not agree with the size of the array (got size $lA for the array and $lI for the indices"))
+    lA == lI || throw(DimensionMismatch("supplied axes do not agree with the size of the array (got size $lA for the array and $lI for the indices"))
     OffsetArray(A, map(indexoffset, inds))
 end
 OffsetArray(A::AbstractArray{T,N}, inds::Vararg{AbstractUnitRange,N}) where {T,N} =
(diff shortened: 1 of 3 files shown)


src/OffsetArrays.jl
```diff
@@ -84,8 +84,26 @@ function Base.similar(A::AbstractArray, ::Type{T}, inds::Tuple{UnitRange,Vararg{
...
+    OffsetArray(T(map(length, shape)), map(indexoffset, shape))
+# Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:Array} =
+#     OffsetArray(T(map(length, shape)), map(indexoffset, shape))
+# Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:BitArray} =
+#     OffsetArray(T(map(length, shape)), map(indexoffset, shape))
+
+# Broadcasting interface
+@static if VERSION > v"0.7.0-DEV.2638" # https://github.com/JuliaLang/julia/pull/23939
+    struct OffsetArrayStyle <: Broadcast.BroadcastStyle end
+    Broadcast.BroadcastStyle(::Type{<:OffsetArray}) = OffsetArrayStyle()
```
timholy: I think you can save several lines with
```julia
Broadcast.BroadcastStyle(::Type{<:OffsetArray}) = Broadcast.ArrayStyle{OffsetArray}()
```
You shouldn't need `broadcast_indices` or define the rule with respect to `Scalar`. More importantly, it should also successfully take precedence over, e.g., `Array`.

test/runtests.jl
```diff
@@ -312,12 +312,12 @@ seek(io, 0)
 amin, amax = extrema(parent(A))
 @test clamp.(A, (amax+amin)/2, amax) == OffsetArray(clamp.(parent(A), (amax+amin)/2, amax), indices(A))
 
-@test unique(A, 1) == parent(A)
-@test unique(A, 2) == parent(A)
+# @test unique(A, 1) == parent(A)
```
timholy: Why do these need to be commented out?

test/runtests.jl
```diff
@@ -312,12 +312,12 @@ seek(io, 0)
...
+# @test unique(A, 2) == parent(A)
 v = OffsetArray(rand(8), (-2,))
 @test sort(v) == OffsetArray(sort(parent(v)), v.offsets)
-@test sortrows(A) == OffsetArray(sortrows(parent(A)), A.offsets)
-@test sortcols(A) == OffsetArray(sortcols(parent(A)), A.offsets)
+# @test sortrows(A) == OffsetArray(sortrows(parent(A)), A.offsets)
```
timholy: Here too?

test/runtests.jl
timholy: Oh, sorry, I see you explained.

I'm less worried about the hijacking than you. To me `similar(BitArray, inds)` means "give me something that *acts* like a BitArray but has the following inds." If that's not a BitArray (generally, it will turn out to be a wrapped BitArray), that's fine.

fredrikekre:
Alright, this just needs https://github.com/JuliaLang/Compat.jl/pull/427 and https://github.com/JuliaLang/Compat.jl/pull/429 now, then a new Compat version.

src/OffsetArrays.jl
```diff
@@ -84,8 +84,22 @@ function Base.similar(A::AbstractArray, ::Type{T}, inds::Tuple{UnitRange,Vararg{
...
     OffsetArray(f(map(length, shape)), map(indexoffset, shape))
+Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:OffsetArray} =
+    OffsetArray(T(map(length, shape)), map(indexoffset, shape))
+Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:Array} =
+    OffsetArray(T(uninitialized, map(length, shape)), map(indexoffset, shape))
+Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:BitArray} =
+    OffsetArray(T(uninitialized, map(length, shape)), map(indexoffset, shape))
+
+# Broadcasting interface
+@static if VERSION > v"0.7.0-DEV.2638" # https://github.com/JuliaLang/julia/pull/23939
```
timholy: Do we need this anymore or is it handled by `similar`?

src/OffsetArrays.jl
```diff
@@ -84,8 +84,22 @@ function Base.similar(A::AbstractArray, ::Type{T}, inds::Tuple{UnitRange,Vararg{
...
 
-Base.similar(f::Union{Function,Type}, shape::Tuple{UnitRange,Vararg{UnitRange}}) =
+Base.similar(f::Function, shape::Tuple{UnitRange,Vararg{UnitRange}}) =
     OffsetArray(f(map(length, shape)), map(indexoffset, shape))
+Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:OffsetArray} =
+    OffsetArray(T(map(length, shape)), map(indexoffset, shape))
+Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:Array} =
+    OffsetArray(T(uninitialized, map(length, shape)), map(indexoffset, shape))
+Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:BitArray} =
+    OffsetArray(T(uninitialized, map(length, shape)), map(indexoffset, shape))
```
timholy: Do we need all three of these specializations or would
```julia
Base.similar(::Type{T}, shape::Tuple{UnitRange,Vararg{UnitRange}}) where {T<:AbstractArray} =
    OffsetArray(T(map(length, shape)), map(indexoffset, shape))
```
be good enough? (OffsetArray being the defacto "owner" of `UnitRange` indices.)

fredrikekre: Yes, the are needed for ambiguity resolution with https://github.com/JuliaLang/julia/blob/7d7d81af41448aeaafb2dadedf54e831616c25cf/base/array.jl#L254 and https://github.com/JuliaLang/julia/blob/7d7d81af41448aeaafb2dadedf54e831616c25cf/base/bitarray.jl#L355. And we also need a separate for `OffsetArray`, which does not use the new `uninitialized` singleton for construction. I plan to implement that in a separate PR.

src/OffsetArrays.jl
fredrikekre: No :disappointed: But it was fun to try the new interface :smile:

timholy:
This seems ready to go except for the 0.6 breakage.

fredrikekre:
Yes, for 0.6 this needs a new Compat tag, was waiting for https://github.com/JuliaLang/Compat.jl/pull/428 before tagging.

fredrikekre:
Looks like tests passes on both 0.6 and 0.7 now :)

timholy:
Wonderful, many thanks for seeing this through.

PR #38, opened 2018-01-01
Rename indices to axes

pablosanjose:
OffsetArrays was failing in Julia master because e.g. `Base.checkbounds` now calls `axes` instead of `indices` (following JuliaLang/#25057), whose generic fallback fails for `OffsetArray`s. Here I just renamed the specialised `indices` to `axes` and `_indices` to `_axes` to fix this failure.


fredrikekre:
Included in #37

PR #39, opened 2018-01-03
Use `uninitialized` in constructor

pablosanjose:
Removes deprecation warning in Julia master, and restores lost performance when creating OffsetArrays


fredrikekre:
Included in #37

timholy:
Thanks for the contribution, and sorry not to merge it. I merged #40 instead.

PR #40, opened 2018-01-16
use uninitialized in OffsetArray constructors

fredrikekre:


src/OffsetArrays.jl
```diff
@@ -27,20 +27,47 @@ OffsetArray(A::AbstractArray{T,N}, offsets::NTuple{N,Int}) where {T,N} =
...
 
 # https://github.com/JuliaLang/julia/pull/19989
-Base.@deprecate OffsetArray(::Type{T}, inds::Vararg{UnitRange{Int},N}) where {T,N} OffsetArray{T}(inds)
-Base.@deprecate OffsetVector(::Type{T}, inds::AbstractUnitRange) where {T} OffsetVector{T}(inds)
+@static if isdefined(Base, :Uninitialized)
+    @deprecate OffsetArray(::Type{T}, inds::Vararg{UnitRange{Int},N}) where {T,N} OffsetArray{T}(uninitialized, inds)
+    @deprecate OffsetVector(::Type{T}, inds::AbstractUnitRange) where {T} OffsetVector{T}(uninitialized, inds)
+else
+    @deprecate OffsetArray(::Type{T}, inds::Vararg{UnitRange{Int},N}) where {T,N} OffsetArray{T}(inds)
+    @deprecate OffsetVector(::Type{T}, inds::AbstractUnitRange) where {T} OffsetVector{T}(inds)
```
fredrikekre: Perhaps these two shouldn't be `@deprecate`. That would mean users get all the deprecation warning at the same time when upgrading to v0.7. Thoughts?

timholy:
Thanks again!

PR #43, opened 2018-04-06
Update in anticipation of deprecating similar(f, ...)

mbauman:
Requires https://github.com/JuliaLang/julia/pull/26733.

This removes the methods dispatching on `similar(f::Function)` in favor of just adding methods to `f` directly.


mbauman:
Now tests pass on 0.6, but fail on Nightly due to the lack of #46.  I'm going to merge this and see what we can do about compatibility with the new axis setup in #46.

PR #46, opened 2018-05-10
Make OffsetArray's axes offset themselves

mbauman:
This is largely an advertisement for JuliaLang/julia#27038 — let's focus the discussion there.  This implements the new scheme proposed there.  Only one failure remains locally with that branch regarding the display of the axes in `summary` since they're now `Slice`s.

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -8,6 +8,9 @@ using Compat: axes, CartesianIndices
 
 export OffsetArray, OffsetVector, @unsafe
 
+## Major change in 0.7: OffsetArray now uses Offset axes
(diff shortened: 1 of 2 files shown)


Evizero:
This breaks 0.6 support, right?

mbauman:
As it stands, yes.  It's not a huge diff so we could potentially fence it off with `if`s, or we could make the next release v0.7-only. I've not been a contributor or maintainer of this package so I'll defer this choice to the folks here.

mbauman:
Ok, I've added back compatibility with 0.6.  It is rather remarkable how small of a change is required to support both 0.6 and 0.7, especially given that the behavior of axes is different between the two versions.

PR #48, opened 2018-06-26
Update size/length workaround recommendation

mbauman:
I'm very tempted to just remove the training wheels altogether, but let's tag a version of OffsetArrays like this first and then we can remove them in the future.


timholy:
Yes, time for the training wheels to come off. I need to get back to https://github.com/JuliaLang/julia/pull/24899 pronto. (Help appreciated, of course!!)

PR #50, opened 2018-07-02
Require julia 0.7-alpha

timholy:
This fixes some new deprecations (e.g., `mapslices`), but mostly it removes version cruft and commits to the new Julia.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ os:
     - linux
 
 julia:
-    - 0.6
     - nightly
 
 notifications:
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,1 @@
-julia 0.6
-Compat 0.45
+julia 0.7.0-beta.73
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -4,7 +4,6 @@ os:
     - linux
 
 julia:
-    - 0.6
```
fredrikekre: Add `- 0.7` instead? Or does this require a newer version than `0.7-beta`?

timholy: No, I just hadn't realized we could say 0.7 yet. Will fix.

fredrikekre: Yea, giving `- 0.7` will use the latest in the `0.7` release series, currently `beta` but it will automatically point to `rc`s and then finally to the release.

.travis.yml
```diff
@@ -4,8 +4,7 @@ os:
...
-    - nightly
```
fredrikekre: Perhaps keep `- nightly` around too, since that is closer to `0.7` compared to `beta`?

timholy: That was interesting---it failed due to mapslices kwds. I upped the minimum julia requirement (it's only for the tests, but still...) and went back to insisting on `nightly`. We can update this later, it's not worth sweating over now.

timholy:
I'll merge when this passes.

PR #54, opened 2018-08-07
Fix final 0.7 deprecations and enable 0.7 CI

mbauman:


mbauman:
The default travis script is currently broken on the 1.0 nightlies and always "passes," but I've now enabled AppVeyor which will hopefully give us the right answer.

PR #59, opened 2018-11-13
Add OffsetArray/OffsetVector constructors for missing and nothing initialization

schmrlng:
Inspired by https://github.com/JuliaLang/julia/blob/9eda36ffa1aa96ddf0ecac3add1144fafc4f9cc9/base/sysimg.jl#L159. I think this implementation with the union type `ArrayInitializer` is a little more compact and should be exactly equivalent to explicitly writing out cases for `UndefInitializer`, `Nothing`, and `Missing` as in the base implementation above, but I could be wrong about the "exactly equivalent" statement -- let me know if you'd prefer the "unrolled" version.


schmrlng:
Bump -- thanks!

timholy:
Thanks!

PR #61, opened 2018-12-10
Update README.md

MasonProtter:
I know there's not anything particularly new here, but I thought it'd be good to show an example where you construct an `OffsetArray` from a preexisting array instead of an `undef`. This might be redundant for experienced users but I think it's helpful for newcomers. The relativistic notation was just the first example that came to mind where `0` based indexing was natural. 

It can be changed to something more natural or accessible if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,29 @@ julia> y[-1,-7,-128,-5,-1,-3,-2,-1] += 5
 19.0
 ```
 
+## Example: Relativistic Notation
+```julia
+julia> using OffsetArrays
+
+julia> r = [:x, :y, :z];
+
+julia> x = OffsetVector([:t, r...], 0:3)
+OffsetArray(::Array{Symbol,1}, 0:3) with eltype Symbol with indices 0:3:
+ :t
+ :x
+ :y
+ :z
+
+julia> x[0]
+:t
+
+julia> x[1:3]
+3-element Array{Symbol,1}:
+ :x
+ :y
+ :z
+```
+
 ## Notes on supporting OffsetArrays
 
 Julia supports generic programming with arrays that doesn't require you to assume that indices start with 1, see the [documentation](http://docs.julialang.org/en/latest/devdocs/offset-arrays/).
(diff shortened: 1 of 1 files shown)


timholy:
Hmm, yes, the current README is indeed pretty anemic. :+1: 

Why do you insert `:t` in the constructor instead of prior? (Just thinking we might as well make this pretty simple.) Also, since I imagine most people will be interested in numeric arrays, what about something like

```julia
η = 0.1
v1 = [sinh(η), cosh(η), 0, 0]
v0 = OffsetArray(v1, 0:3)
```

just to separate the values of the array from setting the indexing?

I don't feel attached to these suggestions, just wondering. And your version might have some advantages, so I'd be happy to merge this as-is.

MasonProtter:
I did `[:t, [:x, :y, :z]...]` because I thought might highlight that I’m interested in a spatial, one-based vector `r = [:x, :y, :z]` and a zero-based spacetime vector `x`. I like this because it emphasizes that 0-based and 1-based arrays might be useful for two related quantities in *one* problem. 

Perhaps it’d e good to have the symbolic example as well as the numeric example? Or I can just switch to your numeric example.

MasonProtter:
I've added an exmple where one may want to express the coefficients of the Laurent polynomial 
```
6/x + 5 - 2 * x + 3 * x^2 + x^3
```
as a `-1`-based array, since if one counts from `-1`, the `nth` coefficient corresponds to the `nth` power of `x`. This has the advantage of introducing the `eachindex` function which is very useful when working with any AbstractArray.

README.md
```diff
@@ -20,6 +20,56 @@ julia> y[-1,-7,-128,-5,-1,-3,-2,-1] += 5
...
+ :z
+```
+
+## Example: Polynomials
+Suppose one wants to represent the Laurent polynomial
+```
+6/x + 5 - 2*x + 3*x^2 + x^3
+```
+in julia. The coefficients of this polynomial are a naturally `-1` based list, since the `n`th element of the list 
+(counting from `-1`) `6, 5, -2, 3, 1` is the coefficient corresponding to the `n`th of `x`. This Laurent polynomial can be evaluated at say `x = 2` as follows.
```
timholy: `n`th *power* of `x`

timholy:
Test failures addressed in #62

README.md
MasonProtter: Good catch! Fixed.

timholy:
Thanks for an excellent and much-needed enhancement!

PR #62, opened 2018-12-11
Support IdentityUnitRange

timholy:
Tests were breaking on nightly, see e.g., https://travis-ci.org/JuliaArrays/OffsetArrays.jl/jobs/466269867.

CC @mbauman

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -3,6 +3,11 @@ VERSION < v"0.7.0-beta2.199" && __precompile__()
 module OffsetArrays
 
 using Base: Indices, tail, @propagate_inbounds
+@static if !isdefined(Base, :IdentityUnitRange)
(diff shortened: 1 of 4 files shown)


mbauman:
Oh bless you. I was just trying to find time in my schedule to do this myself to fix PkgEval in preparation for a 1.1 RC. Turns out this fix is smaller than I expected.

Thank you!

src/OffsetArrays.jl
```diff
@@ -3,6 +3,11 @@ VERSION < v"0.7.0-beta2.199" && __precompile__()
 module OffsetArrays
 
 using Base: Indices, tail, @propagate_inbounds
+@static if !isdefined(Base, :IdentityUnitRange)
+    const IdentityUnitRange = Base.Slice
```
mbauman: I really like how this is now a defined `OffsetArrays` name. Nice.

PR #65, opened 2019-01-17
Add a docstring to the constructor.

tpapp:


tpapp:
I guess it would make sense to document the constructor with `offsets`, too, should I do that?

tpapp:
Now that I revisit this, I am no longer sure what the constructor with indexes does. I thought it specifies the *indices that the result will have*, but then `map(indexoffset, inds)` is not the right function --- consider the pre-#67 behavior (note axes):
```julia
julia> O1 =OffsetArray(ones(3, 3), -1:1, -2:0)
OffsetArray(::Array{Float64,2}, -1:1, -2:0) with eltype Float64 with indices -1:1×-2:0:
 1.0  1.0  1.0
 1.0  1.0  1.0
 1.0  1.0  1.0

julia> O2 =OffsetArray(O1, -1:1, -2:0)
OffsetArray(OffsetArray(::Array{Float64,2}, -1:1, -2:0), -3:-1, -5:-3) with eltype Float64 with indices -3:-1×-5:-3:
 1.0  1.0  1.0
 1.0  1.0  1.0
 1.0  1.0  1.0
```

timholy:
Ooh, you are right. The behavior you achieved in #67 is the one we want to keep, even if it means breaking stuff.

timholy:
The necessary breaking changes were made in #90 and this was added as 2b775200667e30a82f601b4f8008f6669df31d4c

PR #66, opened 2019-01-17
Add no_offset_view

tpapp:
Add a function to return a view with 1-based indexing. Cf [discussion](https://discourse.julialang.org/t/lift-and-wrap-array-with-custom-indexes/19436/11).

I thought this package would be the natural place for this, since it is implemented using `OffsetArray`, but feel free to reject if you don't think it fits into the scope of this package.

Suggestions about the name are welcome.

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -193,4 +193,39 @@ printindices(io::IO, ind1) = print(io, _unslice(ind1))
 _unslice(x) = x
 _unslice(x::IdentityUnitRange) = x.indices
 
+"""
+    no_offset_view(A)
+
+Return an `AbstractArray` that shares structure and has the same type and size as the
+argument, but has 1-based indexing. May just return the argument when applicable. Not
+exported.
+
+The default implementation uses `OffsetArrays`, but other types should use something more
+specific to remove a level of indirection when applicable.
+
+```jldoctest
+julia> O = OffsetArray(A, 0:1, -1:1)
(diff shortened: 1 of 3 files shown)


src/OffsetArrays.jl
```diff
@@ -193,4 +193,39 @@ printindices(io::IO, ind1) = print(io, _unslice(ind1))
...
+
+julia> A
+2×3 Array{Int64,2}:
+ -9  3  5
+  2  4  6
+```
+"""
+function no_offset_view(A::AbstractArray)
+    if Base.has_offset_axes(A)
+        OffsetArray(A, 1 .+ size(A))
```
timholy: This won't work in general:
```julia
julia> using CatIndices, OffsetArrays

julia> v = BidirectionalVector([1,2,3])
BidirectionalVector{Int64} with indices CatIndices.URange(1,3):
 1
 2
 3

julia> pushfirst!(v, 0)
BidirectionalVector{Int64} with indices CatIndices.URange(0,3):
 0
 1
 2
 3

julia> OffsetArrays.no_offset_view(v)
OffsetArray(::BidirectionalVector{Int64}, 5:8) with eltype Int64 with indices 5:8:
 0
 1
 2
 3
```

```suggestion
        OffsetArray(A, map(r->1-first(r), axes(A)))
```

tpapp: Thanks, I will add a test for this.

timholy:
(If you need to rebase there is no need to worry about "credit" for me on that tiny change, just do whatever is easiest.)

tpapp:
Thanks, just resolved the conflict.

timholy:
Thanks, very nice work!

PR #78, opened 2019-06-05
Define some modifying functions for OffsetVector

fredrikekre:
I had these in my own code, might as well upstream them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/Manifest.toml
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,3 @@ julia:
 
 notifications:
     email: false
-
-after_success:
-  - julia -e 'Pkg.init(); run(`ln -s $(pwd()) $(Pkg.dir("OffsetArrays"))`); Pkg.pin("OffsetArrays"); Pkg.resolve()'
-  - julia -e 'using OffsetArrays; @assert isdefined(:OffsetArrays); @assert typeof(OffsetArrays) === Module'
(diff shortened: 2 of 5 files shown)


timholy:
`push!` and `pop!` seem fine. With `pushfirst!`/`popfirst!`, my worry is that the natural thing to do would be to mutate the initial index, but we can't do that without making the wrapper mutable. (See https://github.com/JuliaArrays/CatIndices.jl/blob/master/src/bidirectional.jl.) If we accepted those and then someday got an API for mutating immutables and then decide we want to change how `pushfirst!` works, it becomes non-deprecatable breaking behavior. Of course we could do that (e.g., OffsetArrays v2.0.0), so that's not a killer argument. Nevertheless, I'm tempted to leave those out of this. Thoughts?

fredrikekre:
> With pushfirst!/popfirst!, my worry is that the natural thing to do would be to mutate the initial index

Okay, that is not what `pushfirst!(::Vector)` does though?

timholy:
Right, `pushfirst!` on a vector does not mutate the initial index. So we could say that this is just how `pushfirst!` works. I could agree to that.

OTOH Vector lies in that fuzzy world spanning both views of 1-d arrays, either "I'm just a list of elements" and "I'm an associative container with integer keys." `pushfirst!` seems to belong to the first world, but most uses of OffsetArrays is probably in that second world.

fredrikekre:
Alright, I dropped `(push|pop)first!` for now at least.

fredrikekre:
bump.

Issue #82, opened 2019-10-21
Reshape with Colon doesn't work, can segfault


rapus95:
The reshape call with a Colon doesn't seem to be collected by OffsetArrays, thus due to another bug in Base, one can get segfaults.
See /JuliaLang/julia/issues/33603 and /JuliaLang/julia/issues/33614

timholy:
Fixed by #84

PR #84, opened 2019-11-17
Fix issue #82: Reshape with Colon doesn't work, can segfault

mikan2087:
Fix issue #82.


timholy:
Thanks @mikan2087. Can I ask for a test? Using the MWE from https://github.com/JuliaLang/julia/issues/33614 would seem sufficient.

mikan2087:
@timholy Thank you for reviewing. I added some test cases.

timholy:
Thanks!

PR #90, opened 2019-12-15
Redesign around IdOffsetRange

timholy:
This is breaking and might be a good excuse to go to 1.0. Fixes the observation in https://github.com/JuliaArrays/OffsetArrays.jl/pull/65#issuecomment-457181268 even for other array types that offset their axes (see [this test](https://github.com/JuliaArrays/OffsetArrays.jl/blob/f383ae0d0b139a44e86cc1111470390ddfb83f2e/test/runtests.jl#L60-L65)).

The motivation is discussed in more detail in https://discourse.julialang.org/t/why-is-there-a-performance-hit-on-broadcasting-with-offsetarrays/32194. There's something pretty satisfying, though, about offsetting the axes by a "lazy" mechanism
https://github.com/JuliaArrays/OffsetArrays.jl/blob/f383ae0d0b139a44e86cc1111470390ddfb83f2e/src/OffsetArrays.jl#L18-L21
https://github.com/JuliaArrays/OffsetArrays.jl/blob/f383ae0d0b139a44e86cc1111470390ddfb83f2e/src/OffsetArrays.jl#L41-L42
because it makes it more possible to preserve any custom behavior of the parent array's axes.

CC @mbauman

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "OffsetArrays"
 uuid = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
-version = "0.11.4"
+version = "1.0.0"
 
 [compat]
 julia = "0.7, 1"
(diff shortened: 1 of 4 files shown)


timholy:
Test failure on julia-1.0 might be due to the absence of https://github.com/JuliaLang/julia/pull/28941? We could either add a bound on the Julia version or figure out a workaround.

PR #93, opened 2020-01-03
Preliminary support for matrix multiplication.

benninkrs:
Preliminary support for matrix multiplication.
- New methods for * for combinations of OffsetArray and AbstractArray
  Unfortunately this introduces method ambiguities with Base.
- As an alternative, tried defining a new promote_rule, but somehow it doesn't get invoked.

New convenience constructors
- Wrap any (non-offset) array as an OffsetArray with standard indices
- Remove layer of indirection when constructing with explicit offsets

Cosmetic edits to constructor section for improved readability

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -3,32 +3,43 @@ VERSION < v"0.7.0-beta2.199" && __precompile__()
 module OffsetArrays
 
 using Base: Indices, tail, @propagate_inbounds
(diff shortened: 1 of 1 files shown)


benninkrs:
@timholy  As you can see, I tried a couple things, neither of which worked as well as I'd hoped.  If you have advice either on how to resolve the method ambiguities, or on how to make a `promote_rule` get invoked, I can try to improve it.

PR #99, opened 2020-02-14
Fix unnecessary nested `OffsetArray`

goretkin:
Before this PR, `y1` is`OffsetArray(OffsetArray(...`

```julia
y = OffsetArray(-2:5, -2:5)
y1 = OffsetArray(y, +2)
```

Since the representation of `OffsetArray` just the parent and the offsets, I thought it makes most sense to define this un-nesting at that "base" level. Then any other constructor, e.g. using ranges, benefits.


johnnychen94:
FYI, this change also fixes an "issue" that `parent(y1)` isn't a plain array.

timholy:
Very nice, thanks!

Issue #100, opened 2020-02-27
view != getindex, with IdentityUnitRange


mcabbott:
I got some strange results copying a piece into a larger array, and if I understand right it looks like a problem with how `view` deals with `IdentityUnitRange`. 
```julia
julia> out = reshape(1:15, 5,3) .+ 0; 

julia> item1 = 101:100:301;

julia> out[axes(item1)..., 1] .= item1; 

julia> out 
5×3 Array{Int64,2}:
 101   6  11
 201   7  12
 301   8  13
   4   9  14
   5  10  15

julia> item = OffsetArray(102:100:302, +2)
102:100:302 with indices 3:5

julia> out[axes(item)..., 1] # looks correct  
3-element OffsetArray(::Array{Int64,1}, 3:5) with eltype Int64 with indices 3:5:
 301
   4
   5

julia> view(out, axes(item)..., 1) # wrong? 
3-element view(::Array{Int64,2}, :, 1) with eltype Int64 with indices 3:5:
 5
 6
 7

julia> out[axes(item)..., 2] .= item;

julia> copyto!(view(out, axes(item)..., 2), item); # same effect

julia> out # values of item are in the wrong places!
5×3 Array{Int64,2}:
 101    6  202
 201    7  302
 301    8   13
   4    9   14
   5  102   15

(@v1.5) pkg> st OffsetArrays
Status `~/.julia/environments/v1.5/Project.toml`
  [6fe1bfb0] OffsetArrays v0.11.4
```

johnnychen94:
If you're using OffsetArrays v1 then you would get the right results(i.e., `[301, 4, 5]`) here :P

johnnychen94:
If you wrap it with `CartesianIndices`, it works as expected...

```julia
julia> view(out, CartesianIndices(axes(item)), 1)
3-element view(::Array{Int64,2}, [3, 4, 5], 1) with eltype Int64:
 301
   4
   5

julia> to_indices(item, (axes(item)..., 1))
(3:5, 1)

julia> to_indices(item, (CartesianIndices(axes(item)), 1))
(CartesianIndex{1}[CartesianIndex(3,), CartesianIndex(4,), CartesianIndex(5,)], 1)
```

mcabbott:
Thanks, that's a useful work-around, which I would never have thought to try!

johnnychen94:
update with another result:

```julia
julia> ind = OffsetArrays.IdOffsetRange(3:5, 0)
[...]

PR #101, opened 2020-03-05
Corrects offset1 computing of SubArray for IdOffsetRange

johnnychen94:
I believe this patch somehow fixes issue #100 but it needs careful review since I might misunderstand the design of SubArray and IdOffsetRange...

~issue #100 seems only happens when it's a linear index~

P.S. Tests labeled with `issue #100` were broken without this patch

cc: @mcabbott

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -128,12 +128,9 @@ offset_coerce(::Type{I}, r::AbstractUnitRange) where I<:AbstractUnitRange{T} whe
 @inline Base.axes1(r::IdOffsetRange) = IdOffsetRange(Base.axes1(r.parent), r.offset)
 @inline Base.unsafe_indices(r::IdOffsetRange) = (r,)
 @inline Base.length(r::IdOffsetRange) = length(r.parent)
-@inline function Base.compute_linindex(f, s, IP::Tuple, I::Tuple{IdOffsetRange, Vararg{Any}})
-    # issue 100
-    # IdOffsetRange doesn't contributes to offsets before the construction of SubArray
-    Δi =  1 - first(IP[1])
-    Base.compute_linindex(f + Δi*s, s*Base.unsafe_length(IP[1]), Base.tail(IP), Base.tail(I))
-end
+# issue 100: IdOffsetRange as another index-preserving case shouldn't comtribute offsets
+@inline Base.compute_offset1(parent, stride1::Integer, dims::Tuple{Int}, inds::Tuple{IdOffsetRange}, I::Tuple) =
+    Base.compute_linindex(parent, I) - stride1*first(axes(parent, dims[1]))
 
 @inline function Base.iterate(r::IdOffsetRange)
     ret = iterate(r.parent)
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -128,6 +128,12 @@ offset_coerce(::Type{I}, r::AbstractUnitRange) where I<:AbstractUnitRange{T} whe
 @inline Base.axes1(r::IdOffsetRange) = IdOffsetRange(Base.axes1(r.parent), r.offset)
 @inline Base.unsafe_indices(r::IdOffsetRange) = (r,)
 @inline Base.length(r::IdOffsetRange) = length(r.parent)
+@inline function Base.compute_linindex(f, s, IP::Tuple, I::Tuple{IdOffsetRange, Vararg{Any}})
+    # issue 100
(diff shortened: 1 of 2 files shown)


johnnychen94:
Hmmm. the failures in the nightly test belong to another issue.

johnnychen94:
https://github.com/JuliaLang/julia/pull/34049 break the test here. Am investigating

timholy:
Really amazing work! I'm super-impressed.

My inclination would be to mimic https://github.com/JuliaLang/julia/blob/4d9a3349170d9beea8584cc9a730d37609280ac3/base/subarray.jl#L354-L355, with the thinking that `IdOffsetRange` is another index-preserving case. Care to play with that?

johnnychen94:
Sounds like I'm heading to the right direction

It's a very good exercise for me. Am trying

johnnychen94:
Before I dig deeper and confuse myself, is there any rule that says linear index for OffsetArray should start from 1?

```julia
A = zeros(-3:1, -2:0)
I = axes(A, 1)
A[I] # BoundsError
```

timholy:
Those docs immediately go on to say

> By this definition, 1-dimensional arrays always use Cartesian indexing with the array's native indices.

johnnychen94:
"Obviously" this PR doesn't fix all the things.

```julia
julia> out = zeros(-2:1, -3:-1)
4×3 OffsetArray(::Array{Float64,2}, -2:1, -3:-1) with eltype Float64 with indices -2:1×-3:-1:
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

julia> out[1] = 1
1

julia> ind = axes(out, 1)
-2:1

julia> SubArray(out, (ind, 1)) # it should raise a BoundsError but it doesn't
4-element view(OffsetArray(::Array{Float64,2}, -2:1, -3:-1), -2:1, 1) with eltype Float64 with indices -2:1:
 2.3851344346e-314
 2.385134482e-314
 2.3851346243e-314
 2.3851333437e-314
```

fredrikekre:
I don't think the `SubArray` constructor is meant to be used like that, it completely skips argument checking (presumably for efficiency reasons). You should probably use `view` (or `@view`), which indeed throws a `BoundsError`:
```
julia> view(out, ind, 1)
ERROR: BoundsError: attempt to access 4×3 OffsetArray(::Array{Float64,2}, -2:1, -3:-1) with eltype Float64 with indices -2:1×-3:-1 at index [-2:1, 1]
```

timholy:
Test failure on master is not your fault, so we can merge this and fix it in another PR.

However...with apologies I'm now wondering if this should be fixed in Base (for older Julia versions we will need this, so this PR should be merged). The issue is this: if we fix it here, then other array types that define ranges with axes starting at values other than 1 will have to make the same fix.

Looking at https://github.com/JuliaLang/julia/blob/4d9a3349170d9beea8584cc9a730d37609280ac3/base/subarray.jl#L354-L355, now I'm noting that the method below it is identical in cases where `first(inds) == 1`. Since that's the default for ranges anyway, should we just delete the method below it?

The hard part will be writing a test. If you're interested in tackling this, you could copy much of `OffsetArrays/src/axes.jl` to `test/testhelpers/OffsetArrays.jl` if you wish. Or define something like
```julia
# Use a range type not defined in Base
struct StrangeRange <: AbstractUnitRange{Int}
    stop::Int
end
Base.first(r::StrangeRange) = -r.stop
Base.last(r::StrangeRange)  =  r.stop
Base.axes(r::StrangeRange)  = (0:2*r.stop,)
...
```
Note this is a case where the values and the axes are deliberately misaligned AND the first element of neither is 1. If that doesn't shake out bugs...

If you don't want to tackle Base, just let me know and I can do it.

johnnychen94:
> Looking at JuliaLang/julia:base/subarray.jl@4d9a334#L354-L355, now I'm noting that the method below it is identical in cases where first(inds) == 1. Since that's the default for ranges anyway, should we just delete the method below it?

The benchmark says we'd better not do so...

```julia
julia> A = rand(5, 5);

julia> I = Base.IdentityUnitRange(3:5)
Base.IdentityUnitRange(3:5)

julia> @btime Base.compute_offset1(A, 1, (I, 1));
  120.822 ns (1 allocation: 32 bytes)

julia> @btime Base.compute_offset1(A, 1, (3:5, 1));
  30.544 ns (0 allocations: 0 bytes)
```

PR #102, opened 2020-03-06
"fix" nightly test

johnnychen94:
Reading through the discussion it's more like an upstream bug fix. We could remove these tests, or conditionally run them in different versions...

cref: https://github.com/JuliaLang/julia/pull/34049


timholy:
Very nice!

PR #103, opened 2020-03-07
add parent(r::IdOffsetRange)

johnnychen94:
unless there's a reason not to...

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -124,6 +124,7 @@ end
 offset_coerce(::Type{I}, r::AbstractUnitRange) where I<:AbstractUnitRange{T} where T =
     convert(I, r), 0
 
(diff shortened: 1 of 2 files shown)


timholy:
Good idea, but needs a test.

fredrikekre:
I don't think I have ever seen a good usage of `parent`. What can you use it (in generic code) for?

johnnychen94:
> I don't think I have ever seen a good usage of parent. What can you use it (in generic code) for?

Since in many cases we're using `parent(A)` and `A.parent` interchangeably, I think adding this can be useful for convenience and for consistency. At least for this PR, I meant this.

I'm totally fine not adding this, but it's likely that sometimes later someone would find this missing and try to add it again.

PR #115, opened 2020-06-11
Reuse `similar` for `undef` initialization

goretkin:
I didn't check carefully, but I expect this is equivalent, and it looks clearer to me.

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -126,15 +126,15 @@ function Base.similar(::Type{T}, shape::Tuple{OffsetAxis,Vararg{OffsetAxis}}) wh
 end
 
 Base.fill(v, inds::NTuple{N, Union{Integer, AbstractUnitRange}}) where {N} =
-    fill!(OffsetArray(Array{typeof(v), N}(undef, map(indexlength, inds)), map(indexoffset, inds)), v)
+    fill!(similar(Array{typeof(v)}, inds), v)
 Base.zeros(::Type{T}, inds::NTuple{N, Union{Integer, AbstractUnitRange}}) where {T, N} =
-    fill!(OffsetArray(Array{T, N}(undef, map(indexlength, inds)), map(indexoffset, inds)), zero(T))
+    fill!(similar(Array{T}, inds), zero(T))
 Base.ones(::Type{T}, inds::NTuple{N, Union{Integer, AbstractUnitRange}}) where {T, N} =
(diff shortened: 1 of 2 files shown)


goretkin:
Thanks for taking a look. I will take a look at the extra `fill`.

I don't immediately understand your first comment. I think you're talking about how this package defines methods for both `similar(A::AbstractArray, ...)`  and `similar(::Type{T}, ...) where {T<:AbstractArray}`, and that possibly there's a fallback through base so that there is just one implementation? In any case, I will take a look to see what can be done, but also wanted to ask for clarification.

src/OffsetArrays.jl
```diff
@@ -126,15 +126,15 @@ function Base.similar(::Type{T}, shape::Tuple{OffsetAxis,Vararg{OffsetAxis}}) wh
 end
 
 Base.fill(v, inds::NTuple{N, Union{Integer, AbstractUnitRange}}) where {N} =
-    fill!(OffsetArray(Array{typeof(v), N}(undef, map(indexlength, inds)), map(indexoffset, inds)), v)
+    fill!(similar(Array{typeof(v)}, inds), v)
```
johnnychen94, line 129: > I don't immediately understand your first comment.

Sorry for the inaccuracy, I was saying `similar` here actually calls this method:
 
https://github.com/JuliaArrays/OffsetArrays.jl/blob/7294f3dcab6e6117b15eea3b98ce2c2175d08194/src/OffsetArrays.jl#L123-L126 

and thus is equivalent to the previous implementation, if I intuited it correctly.

goretkin:
I was adding a few more tests, and I see that there are some places where `===` is used, which is good since `1:3 == Base.OneTo(3)`, and some places want to test that.

But I also see many cases of e.g. `a == IdOffsetRange(3:5)` being used, and I don't see the point of doing that instead of `a == 3:5`, unless the test is really intended to check for `===`, in which case in at least a few places it should be `a === OffsetArrays.IdOffsetRange(Base.OneTo(3), 2)`.

goretkin:
Same with `IdentityUnitRange`. I think I'll open an issue for this.

src/OffsetArrays.jl
goretkin: I think I just realized you might be making an observation, and not making a suggestion or a request. Is that right? Sorry again for still not understanding. :-D

johnnychen94:
Thanks for the work @goretkin !

PR #135, opened 2020-08-29
use Base implementation for compute_offset1

johnnychen94:
This continues #134

With https://github.com/JuliaLang/julia/pull/37204 merged, this patch is no longer needed for the latest Julia version.

cc: @jishnub


johnnychen94:
Thank you @jishnub for digging into the details and reporting the bug upstream

PR #137, opened 2020-09-03
WIP: Implement insert! and deleteat! for OffsetVectors

garrison:
I consider this to be incomplete.  I have not, for instance, thought carefully about whether I am referring to the `axes` in the most appropriate way, but hey, all tests pass!  Also, ideally the offset indices will be used in the exception message when the index is out of range; right now, the indices of the parent array are instead provided, and this can be confusing.

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -151,7 +151,7 @@ Base.falses(inds::NTuple{N, Union{Integer, AbstractUnitRange}}) where {N} =
 #   Δi = i - first(r)
 #   i′ = first(r.parent) + Δi
 # and one obtains the result below.
-parentindex(r::IdOffsetRange, i) = i - r.offset
+parentindex(r::IdOffsetRange, i) = i .- r.offset
 
 @inline function Base.getindex(A::OffsetArray{T,N}, I::Vararg{Int,N}) where {T,N}
     @boundscheck checkbounds(A, I...)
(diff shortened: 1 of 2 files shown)


garrison:
Just found https://github.com/JuliaArrays/OffsetArrays.jl/pull/55#issuecomment-438525109, which did not turn up in my earlier repository search.  I still believe implementing `insert!` and `deleteat!` is valuable, but obviously it deserves some discussion first.  I'll open an issue shortly to discuss the related question of whether to implement `pushfirst!` and `popfirst!`.

timholy:
As you can tell from your readings, I'm mixed on this. If I had to take a stance, it would be "against by a hair." Here's why:
- `pushfirst!` really seems like it should grow in the "opposite direction", meaning `firstindex(a)` gets decremented rather than `lastindex(a)` being incremented
- insert!(a, firstindex(a), items)` should behave the same as `pushfirst!` if `items` is of length 1.

MasonProtter:
@timholy  What about `deleteat!`? I think that's a useful function and it's required for things like `filter!` and would have been useful test cases for two PRs I recently wrote: https://github.com/JuliaLang/julia/pull/39528 and https://github.com/Vexatos/CircularArrays.jl/pull/12

timholy:
We can have this if there's enough enthusiasm. I just think that the consensus has never been clear. Lots of people initially think "why isn't that implemented? I'll add it!" and then once the concerns are raised most of us seem to think that it's not obvious we want it. If someone makes a sufficiently clear case and is willing to document the behavior carefully, we can do things like this. But the principle of "do no harm" holds me back from adding this until someone really pushes a strong argument.

PR #138, opened 2020-09-03
Implement append!

garrison:
This seems relatively non-controversial, since `pop!` and `push!` already exist.


johnnychen94 (approved):
LGTM

timholy:
Thanks!

PR #140, opened 2020-09-08
meta project quality test with Aqua

johnnychen94:
How ambiguity is computed is changed in https://github.com/JuliaLang/julia/pull/36962

For the purpose of ambiguity check for this package, it would be pretty much sufficient to only check that we don't contribute more ambiguities to Base and Core.

Aqua is used, instead, to free us from manually maintaining these meta project quality tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,10 +6,11 @@ version = "1.1.3"
 julia = "0.7, 1"
 
 [extras]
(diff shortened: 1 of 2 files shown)


timholy:
I like it! Thanks for introducing me to Aqua.

PR #160, opened 2020-09-23
document OffsetVector and OffsetMatrix

tpapp:


johnnychen94 (approved):
They is actually an initialization version:

```julia
OffsetVector{T}(::Initializer, offset_range)
```

but since this is not documented in `OffsetArray` neither. I think it is okay to get this in.

PR #164, opened 2020-10-09
inline OffsetArray constructors

johnnychen94:
To reduce the overhead

```julia
julia> a = zeros(5,5,5,5);

julia> @btime OffsetArray($a, -2:2, -2:2, -2:2, -2:2);
  3.088 ns (0 allocations: 0 bytes) # this PR; 1.5.2
  2.463 ns (0 allocations: 0 bytes) # this PR; 1.6-dev
  16.180 ns (0 allocations: 0 bytes) # master; 1.5.2
  15.693 ns (0 allocations: 0 bytes) # master; 1.6-dev
```

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -93,7 +93,7 @@ struct OffsetArray{T,N,AA<:AbstractArray{T,N}} <: AbstractArray{T,N}
     parent::AA
     offsets::NTuple{N,Int}
     function OffsetArray{T, N, AA}(parent::AA, offsets::NTuple{N, Int}) where {T, N, AA<:AbstractArray{T,N}}
-        @boundscheck overflow_check.(axes(parent), offsets)
+        overflow_check.(axes(parent), offsets)
         new{T, N, AA}(parent, offsets)
     end
 end
@@ -126,7 +126,7 @@ end
 
 # Tuples of integers are treated as offsets
 # Empty Tuples are handled here
-function OffsetArray(A::AbstractArray, offsets::Tuple{Vararg{Integer}})
+@inline function OffsetArray(A::AbstractArray, offsets::Tuple{Vararg{Integer}})
     _checkindices(A, offsets, "offsets")
     OffsetArray{eltype(A), ndims(A), typeof(A)}(A, offsets)
 end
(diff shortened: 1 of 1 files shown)


timholy (approved):
It doesn't get inlined now due to `length(::UnitRange{Int})` having two checked-arithmetic operations. (See `Base.print_statement_costs` on 1.6.) This seems reasonable, and gives Julia a chance to elide the wrapper in the caller.

src/OffsetArrays.jl
```diff
@@ -93,7 +93,7 @@ struct OffsetArray{T,N,AA<:AbstractArray{T,N}} <: AbstractArray{T,N}
     parent::AA
     offsets::NTuple{N,Int}
     function OffsetArray{T, N, AA}(parent::AA, offsets::NTuple{N, Int}) where {T, N, AA<:AbstractArray{T,N}}
-        @boundscheck overflow_check.(axes(parent), offsets)
+        overflow_check.(axes(parent), offsets)
```
fredrikekre, line 96: It seems like this would allocate an array, is that completely elided?

timholy: Good thought. Really this should be a `foreach` call.

johnnychen94: Hmmm, turns out `map` is faster here.

* map: 3.004ns
* foreach: 5.261 ns

Is there a way to check if a method is properly inlined?

timholy: `foreach` doesn't force-specialize on the function.

> Is there a way to check if a method is properly inlined?

`@code_typed` or Cthulhu

fredrikekre: > I would guess, though, that it's elided. It should be creating a tuple

Yea looks from the benchmarks that it is elided, it's just not immediatly obvious from the code ( I didnt't think about the inputs beeing tuples). `map` is probably fine then (with a comment?).

johnnychen94: > `map` on tuple can be optimized by the compiler and thus not using `foreach` here

This is what I added. Unsure what's the terminology here, "constant folding"?

timholy: You could just say "optimized away." Another option would be [DCE](https://en.wikipedia.org/wiki/Dead_code_elimination).

PR #236, opened 2021-05-15
fix for Iterators.partition on IdOffsetRange

N5N3:
before
```julia
Iterators.partition(IdOffsetRange(500:1001,10000),500)|> collect
2-element Vector{UnitRange{Int64}}:
 500:999
 1000:1001
```
after
```julia
Iterators.partition(IdOffsetRange(500:1001,10000),500)|> collect
2-element Vector{UnitRange{Int64}}:
 10500:10999
 11000:11001
```

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -195,7 +195,7 @@ end
 @inline function Base.iterate(itr::Iterators.PartitionIterator{<:IdOffsetRange}, (s, e))
     s > e && return nothing
     r = min(s + itr.n - 1, e)
-    return @inbounds itr.c[s:r], (r + 1, e)
+    return @inbounds view(itr.c,s:r), (r + 1, e) # use view here, to keep consistancy
 end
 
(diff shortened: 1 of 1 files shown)


N5N3:
Well, I think this should still be a IdOffsetRange ?

johnnychen94:
Without looking in the details, I'm a little bit resistant to this PR because it doesn't look like a generic solution; there're many other iterator types outside and may have similar issues, this patch only fixes one small case.

N5N3:
From this example, any wrapped Iterator assuming that AbstractRange is 1-based indexed might have a similar problem. So you mean it's better to fix it in `Iterators`?

johnnychen94:
> So you mean it's better to fix it in Iterators?

If there's some internal method that we can extend here and so as to support almost all iterators, that would be great. But I don't know if there's such a function. So yes, it's probably better to enhance those iterators directly in their Base implementation.

N5N3:
Well, i'll make a PR

PR #262, opened 2021-09-02
fixes #261

### LINKED ISSUE #261: centered should accept CartesianIndex
Seems logical that:
```julia
OffsetArrays.centered(rand(3,3), CartesianIndex(2,2))
```
would work.

yakir12:


johnnychen94:
@yakir12 A suggestion for future PRs (commits), it's usually more descriptive to have some words to describe the feature and not just issue links. For instance, "Allow CartesianIndex for `centered` (fixes #262)" might be a good candidate. Only having issue numbers requires people to click into the page and check what that is really about.

yakir12:
You're totally right. I'll do that from now on.

PR #268, opened 2021-10-15
Depend on ArrayInterface

Tokazama:
This is an initial attempt to start getting rid of `Requires` for ArrayInterface.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,15 +4,19 @@ version = "1.10.7"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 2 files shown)


timholy:
I think we want ArrayInterface pretty broadly, and if it's going to be loaded by multiple packages then it won't have a direct impact on this package (it will already be loaded by something else).

timholy:
@johnnychen94's concerns about potential divergence for testing Base are quite relevant. For 1.0 compatibility, I'd agree that it's a disadvantage until Julia 1.6 is officially declared LTS. Once that happens, I say we abandon any duty to maintain compat with 1.0.

PR #269, opened 2021-11-03
Stop pirating `mapreduce(f, op)`

Keno:


timholy:
Thanks to both! https://github.com/JuliaRegistries/General/pull/48130

PR #270, opened 2021-12-26
Allow some linear algebra

mcabbott:
This is meant to work with https://github.com/JuliaLang/julia/pull/43552, although I think `mul!` might work without that.

Alternative to #146. Since this does not overload `*`, I think it should not encounter the endless method ambiguities that tends to cause, against Adjoint matrices and other types. In fact I think other packages could overload `generic_matmul!` too in the same way, and each unwrap nicely, so long as they all only dispatch on the output `C`, which is created by `similar`.

This seems to cause one extra allocation and hence is slower than without offsets. I'm not so sure why, I thought `MulAddMul(α,β)` (whose type depends on the values of α,β) was the culprit, but in fact get similar times after avoiding that:
```julia
julia> for N in [3, 10, 100, 1000]
           @show N
           A, B, x = rand(N,N), rand(N,N), rand(N)
           @btime $A * $B
           A, B, x = OffsetArray(A,5,5), OffsetArray(B,5,5), OffsetArray(x,5)
           @btime $A * $B
       end
N = 3
  min 31.407 ns, mean 33.633 ns (1 allocation, 128 bytes)
  min 52.950 ns, mean 56.294 ns (2 allocations, 144 bytes)
N = 10
  min 180.901 ns, mean 200.923 ns (1 allocation, 896 bytes)
  min 209.057 ns, mean 225.422 ns (2 allocations, 912 bytes)
N = 100
  min 9.625 μs, mean 15.880 μs (2 allocations, 78.17 KiB)
  min 9.583 μs, mean 15.396 μs (3 allocations, 78.19 KiB)
N = 1000
  min 7.685 ms, mean 8.350 ms (2 allocations, 7.63 MiB)
  min 7.756 ms, mean 8.336 ms (3 allocations, 7.63 MiB)
```

### DIFF
--- a/src/linearalgebra.jl
+++ b/src/linearalgebra.jl
@@ -1,0 +1,94 @@
+using LinearAlgebra
+using LinearAlgebra: MulAddMul, mul!
+lapack_axes(t::AbstractChar, M::AbstractVecOrMat) = (axes(M, t=='N' ? 1 : 2), axes(M, t=='N' ? 2 : 1))
+
+# The signature of this differs from LinearAlgebra's only on C
+function LinearAlgebra.generic_matvecmul!(C::OffsetVector, tA, A::AbstractVecOrMat, B::AbstractVector,
+                            _add::MulAddMul = MulAddMul())
+
+    mB_axis = Base.axes1(B)
+    mA_axis, nA_axis = lapack_axes(tA, A)
+
+    if mB_axis != nA_axis
+        throw(DimensionMismatch("mul! can't contract axis $(UnitRange(nA_axis)) from A with axes(B) == ($(UnitRange(mB_axis)),)"))
+    end
+    if mA_axis != Base.axes1(C)
+        throw(DimensionMismatch("mul! got axes(C) == ($(UnitRange(Base.axes1(C))),), expected $(UnitRange(mA_axis))"))
+    end
+
+    C1 = no_offset_view(C)
+    A1 = no_offset_view(A)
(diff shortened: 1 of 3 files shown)


src/linearalgebra.jl
```diff
@@ -0,0 +1,113 @@
...
+    A1 = no_offset_view(A)
+    B1 = no_offset_view(B)
+
+   if tA == 'N'
+        if tB == 'N'
+            mul!(C1, A1, B1, alpha, beta)
+        elseif tB == 'T'
+            mul!(C1, A1, transpose(B1), alpha, beta)
+        elseif tB == 'C'
+            mul!(C1, A1, adjoint(B1), alpha, beta)
```
mcabbott, line 79: A bigger change to `LinearAlgebra.generic_matmatmul!` would be to make it keep `adjoint` longer, before introducing `'C'`, etc. Then this nest of conditions could be removed.

It seems an odd design that MulAddMul pushes `α,β` into the type domain (partly) at the same time that it moves transpose/adjoint to values from types. Perhaps https://github.com/JuliaLang/julia/pull/43552 could fix both at the same time.

mcabbott: 816b928 (and https://github.com/JuliaLang/julia/pull/43552/commits/aad052225e06208a979bdc23222a391d469f3343) changes this. It removes the extra allocation above.

The downside is that these methods will not be called for `mul!` with a version of Julia older than  JuliaLang/julia#43552 .

PR #290, opened 2022-05-22
Implement Base.unsafe_wrap for OffsetArrays (pull request)

### LINKED ISSUE #275: Implement Base.unsafe_wrap for OffsetArrays
`Base.unsafe_wrap` is not implemented for `OffsetArrays` so this is not possible.

```julia
julia> p = Libc.malloc(1024)
Ptr{Nothing} @0x0000000003efaf10

julia> unsafe_wrap(OffsetArray{Int,1}, p, (1024,); own = true)
ERROR: StackOverflowError:
Stacktrace:
 [1] unsafe_wrap(Atype::Type, p::Ptr{Nothing}, dims::Tuple{Int64}; own::Bool) (repeats 14529 times)
   @ Base ./pointer.jl:92
 [2] top-level scope
   @ REPL[31]:1
```

I am interested in being able to compose https://github.com/mkitti/ArrayAllocators.jl with OffsetArrays.

mkitti:
Fix #275

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -623,6 +623,25 @@ function Base.replace_in_print_matrix(A::OffsetArray{<:Any,1}, i::Integer, j::In
     Base.replace_in_print_matrix(parent(A), ip, j, s)
 end
 
+@inline function Base.unsafe_wrap(::Union{Type{OffsetArray}, Type{OffsetArray{T}}, Type{OffsetArray{T,N}}, Type{OffsetArray{T1, N} where T1}}, pointer::Ptr{T}, inds::NTuple{N, OffsetAxisKnownLength}; own = false, kw...) where {T,N}
+    _checkindices(N, inds, "indices")
+    AA = Base.unsafe_wrap(Array, pointer, map(_indexlength, inds); own=own)
+    OffsetArray{T, N, typeof(AA)}(AA, map(_indexoffset, inds); kw...)
+end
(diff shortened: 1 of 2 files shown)


src/OffsetArrays.jl
```diff
@@ -623,6 +623,25 @@ function Base.replace_in_print_matrix(A::OffsetArray{<:Any,1}, i::Integer, j::In
     Base.replace_in_print_matrix(parent(A), ip, j, s)
 end
 
+@inline function Base.unsafe_wrap(::Union{Type{OffsetArray}, Type{OffsetArray{T}}, Type{OffsetArray{T,N}}, Type{OffsetArray{T1, N} where T1}}, pointer::Ptr{T}, inds::NTuple{N, OffsetAxisKnownLength}; own = false, kw...) where {T,N}
```
timholy, line 626: If I'm not misreading, this and the next method have identical bodies. Could we call this method `_unsafe_wrap` and then define two methods of `Base.unsafe_wrap` that simply call `_unsafe_wrap`?

timholy (commented):
The implementation is excellent, only one tiny change to request.

src/OffsetArrays.jl
mkitti: Yes, I think we could. Another option would be to reduce each to a single line call to an existing constructor.

timholy:
https://github.com/JuliaRegistries/General/pull/60923

PR #371, opened 2025-03-12
Trying to allow `Int8` as indices

simone-silvestri:
This PR is trying to solve #370. This is just an initial proposal, but I guess there is a lot of work to do still.
I have noticed that
```julia
julia> BigInt(1) !== BigInt(1)
true
```
so I had to manually change some tests when the offsets are `BigInt`s

Any suggestion or comment is more than welcome as I am not very sure about downstream implications of this PR

### DIFF
--- a/src/OffsetArrays.jl
+++ b/src/OffsetArrays.jl
@@ -183,9 +183,6 @@ for FT in (:OffsetArray, :OffsetVector, :OffsetMatrix)
         I = map(+, _offsets(A, parent(A)), offsets)
         $FT(parent(A), I, checkoverflow = false)
     end
-    @eval @inline function $FT(A::OffsetArray, offsets::Tuple{Integer,Vararg{Integer}}; kw...)
-        $FT(A, map(Int, offsets); kw...)
-    end
 
     # In general, indices get converted to AbstractUnitRanges.
     # CartesianIndices{N} get converted to N ranges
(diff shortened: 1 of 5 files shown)


src/OffsetArrays.jl
```diff
@@ -530,7 +547,7 @@ end
 # An OffsetUnitRange{Int} has an equivalent IdOffsetRange with the same values and axes,
 # something similar also holds for OffsetUnitRange{BigInt}
 # We may replace the former with the latter in an indexing operation to obtain a performance boost
-@inline function Base.to_index(r::OffsetUnitRange{<:Union{Int,BigInt}})
+@inline function Base.to_index(r::OffsetUnitRange{<:Union{Integer,BigInt}})
```
giordano: `BigInt <: Integer`
```suggestion
@inline function Base.to_index(r::OffsetUnitRange{<:Integer})
```

giordano (commented):
I'll defer the decision about adding a new type parameter to the maintainers of the package

simone-silvestri:
Would it be ok if I find a way to implement this functionality without introducing a new parameter?

simone-silvestri:
Actually, thinking a bit about it, it seems to be a bit to difficult to achieve without an extra parameter. Another possibility would be to just allow this feature in downstream packages, for example by defining an abstract type `AbstractOffsetArray{T, N}` which could be subtyped and move most of the methods to target `AbstractOffsetArray`. Would that be an option?

jishnub:
Yes, that makes sense

